PR #173, opened 2024-07-22
Run lean CI in the right directory

eric-wieser:
Follows up from #170, which was misconfigured.

If you look at the build logs, you will see that the build runs as expected, though fails on some files that genuinely have errors in them.


amit9oct (approved):

eric-wieser:
I was not expecting you to merge this while there were build errors!

PR #177, opened 2024-07-25
Isabelle & Coq verifications: 2004-2005 & 1991-1995

MichaelHenryJennings:


coq/src/putnam_1991_a2.v
GeorgeTsoukalas, line 1: I don't think this is introducing real matrices, only those over some ring R. Right? I'm not sure if it is actually possible to write this one with the existing support.

coq/src/putnam_1992_b4.v
GeorgeTsoukalas, line 1: Is this asserting something about real polynomials? This may need to use the manual polynomial definition as not sure that (R: numDomainType) refers to the reals?

lean4/src/putnam_1995_a4.lean
```diff
@@ -6,7 +6,7 @@ open Filter Topology Real
 theorem putnam_1995_a4
 (n : ℕ)
 (hn : n > 0)
-(necklace : Fin n → ℤ)
-(hnecklacesum : ∑ i : Fin n, necklace i = n - 1)
-: ∃ cut : Fin n, ∀ k : Fin n, ∑ i : {j : Fin n | j.1 ≤ k}, necklace (cut + i) ≤ k :=
+(necklace : ℕ → ℤ)
+(hnecklacesum : ∑ i ∈ Finset.range n, necklace i = n - 1)
```
eric-wieser, line 10: This is a syntax error!

PR #188, opened 2024-08-01
Fixed Euclidean dist issues w.t.t. Eric Wieser for his suggestion

GeorgeTsoukalas:


lean4/src/putnam_1963_a6.lean
```diff
@@ -4,11 +4,11 @@ open BigOperators
 open Topology Filter
 
 theorem putnam_1963_a6
-(F1 F2 U V A B C D P Q : Fin 2 → ℝ)
+(F1 F2 U V A B C D P Q : EuclideanSpace ℝ (Fin 2))
 (r : ℝ)
-(E : Set (Fin 2 → ℝ) := {H : Fin 2 → ℝ | Euclidean.dist F1 H + Euclidean.dist F2 H = r})
-(M : Fin 2 → ℝ := midpoint ℝ U V)
-(hr : r > Euclidean.dist F1 F2)
+(E : Set (Fin 2 → ℝ) := {H : EuclideanSpace ℝ (Fin 2) | dist F1 H + dist F2 H = r})
```
eric-wieser, line 9: `Fin 2 → ℝ` here is the wrong type, it should also be `EuclideanSpace ℝ (Fin 2)`

lean4/src/putnam_1994_a3.lean
```diff
@@ -3,10 +3,9 @@ open BigOperators
 
 open Filter Topology
 
+-- Note: The formalization here differs slightly from the original problem statement, in that T is the entire triangle, not just the sides. We adopt the former interpretation because it is simpler to state and maintains the difficulty of the problem.
 theorem putnam_1994_a3
-(vec : ℝ → ℝ → (Fin 2 → ℝ))
-(T : Set (Fin 2 → ℝ) := convexHull ℝ {vec 0 0, vec 1 0, vec 0 1})
+(T : Set (EuclideanSpace ℝ (Fin 2)) := convexHull ℝ {![0,0], ![1,0], ![0,1]})
```
eric-wieser, line 8: Strictly you need to launder these `![_, _]` vectors through `WithLP.equiv`, to convert them to the right type. Probably the statement is still correct either way.

lean4/src/putnam_1971_a3.lean
```diff
@@ -9,7 +9,7 @@ theorem putnam_1971_a3
 (habclattice : a.1 = round a.1 ∧ a.2 = round a.2 ∧ b.1 = round b.1 ∧ b.2 = round b.2 ∧ c.1 = round c.1 ∧ c.2 = round c.2)
 (habcneq : a ≠ b ∧ a ≠ c ∧ b ≠ c)
 (hR : R > 0)
-(oncircle : (ℝ × ℝ) → ℝ → (ℝ × ℝ) → Prop := fun C r p => Euclidean.dist p C = r)
+(oncircle : (ℝ × ℝ) → ℝ → (ℝ × ℝ) → Prop := fun C r p => Real.sqrt ((p.1 - C.1)^2 + (p.2 - C.2)^2) = r)
 (hcircle : ∃ C : ℝ × ℝ, oncircle C R a ∧ oncircle C R b ∧ oncircle C R c)
```
eric-wieser, line 13: If you use `WithLP 2 (ℝ × ℝ)` here and elsewhere instead of `ℝ × ℝ`, then `dist` has the right meaning.

lean4/src/putnam_1963_a6.lean
GeorgeTsoukalas: Good catch! Should be fixed in incoming PR.

lean4/src/putnam_1994_a3.lean
GeorgeTsoukalas: Hm, will specify it anyways in the chance it would break in a future version of Lean. Thanks!

lean4/src/putnam_1971_a3.lean
GeorgeTsoukalas: Thanks! For this one, I think I'll change some problems which I come across and leave others as is. Some variation here can help identify weaknesses in methods people end up trying down the line. Statement should be correct either way but different methods might be more capable on one than the other, which should be useful info to have.

PR #189, opened 2024-08-02
Jasper

leejasper851:
1996-1998 Isabelle and Coq fixes, plus some other fixes


coq/src/putnam_1997_b5.v
GeorgeTsoukalas, line 1: Can you modify powm -> tetration for consistency with naming in the Lean version?

GeorgeTsoukalas (approved):

PR #193, opened 2024-08-03
optParam and notational improvements.

GeorgeTsoukalas:


lean4/src/putnam_1962_b6.lean
```diff
@@ -4,11 +4,13 @@ open BigOperators
 open MeasureTheory
 
 theorem putnam_1962_b6
-(π : ℝ := Real.pi)
+(π : ℝ)
+(hπ : π = Real.pi)
```
eric-wieser, line 8: `open scoped Real` makes this unecessary.

GeorgeTsoukalas: Thanks! This was done in 2 problems which should be amended shortly.

PR #197, opened 2024-08-09
1999 to 2001 isabelle and coq fixes

leejasper851:


coq/src/putnam_1999_a2.v
GeorgeTsoukalas, line 1: My expectation is that this problem is false with R : numDomainType. With new mathcomp-analysis dependency (see updated setup script) you can import `reals` from mathcomp and change numDomainType -> realType.

amit9oct (approved):

PR #203, opened 2024-08-16
Fix some minor Lean misformalisations

ocfnash:


GeorgeTsoukalas (commented):
Thank you so much for the PR! We really appreciate the contribution. Good catches with the subtle mistakes, we missed these in our verification. There also appeared to be some issues which probably arose from doing too many global syntactic edits at once. We did the most recent years first so hopefully there are less in the remaining problems. We also were not previously aware of `IsGreatest` and `IsLeast`, there are a fair amount of problems which still spell it out directly. We can get to those, but it might be several weeks as the semester is starting up soon so most of the team will be busy, and I'll be traveling shortly. 

We'll also make some modifications in the corresponding Isabelle & Coq files to match any edits. Will leave this PR until we have made those changes. I also noted in `putnam_2021_b2` that the held out solution is not mentioned in the theorem statement, but don't worry about this we can make this change on our end.

lean4/src/putnam_2021_b2.lean
GeorgeTsoukalas, line 1: In the refactoring `(2 / 3)` is included instead of `putnam_2021_b2_solution`.

ocfnash: Good catch, I've fixed this in 64e63b68cd1804179341d0283b0eadabaf259d56

ocfnash:
> Thank you so much for the PR!
You're very welcome :)

> We also were not previously aware of `IsGreatest` and `IsLeast`, there are a fair amount of problems which still spell it out directly. We can get to those, but it might be several weeks 
Yes I was intending to highlight this. It's good to hear you might use these more widely, but there's no rush here. I think fixing misformalisations is much more urgent.

> We'll also make some modifications in the corresponding Isabelle & Coq files to match any edits. Will leave this PR until we have made those changes.
Of course it's your project so you should certainly act as you see fit. However I'm not convinced that there is benefit to keeping a consistency of misformalisations across the three languages. This is just a batch of the first 10 problems I decided to fix and I should be able to fix many more. However if I have to wait for the Isabelle & Coq changes each time, I may not have further PRs.

GeorgeTsoukalas:
> However I'm not convinced that there is benefit to keeping a consistency of misformalisations across the three languages. This is just a batch of the first 10 problems I decided to fix and I should be able to fix many more. However if I have to wait for the Isabelle & Coq changes each time, I may not have further PRs.

Yeah, on second thought I agree with you. I'll merge and open an issue for the other languages. Let me know if you have any other suggestions!

GeorgeTsoukalas (approved):

PR #205, opened 2024-08-17
Unknown identifier fixes (Lean).

MichaelHenryJennings:


eric-wieser:
Nice, this closes #202 I think

lean4/src/putnam_2017_b2.lean
```diff
@@ -11,7 +11,7 @@ theorem putnam_2017_b2
 (S : ℤ → ℕ → ℤ)
 (p : ℤ → ℕ → Prop)
 (q : ℤ → Prop)
-(hS : S = fun a k ↦ ∑ i : Fin k, a + i)
+(hS : S = fun (a : ℤ) k ↦ ∑ i in Finset.range k, (a + i))
```
eric-wieser, line 14: You shouldn't need the type ascription here

MichaelHenryJennings: When I remove it, I get the error:
> type mismatch
  fun a k => ∑ i ∈ Finset.range k, (a + i)
has type
  ℕ → ℕ → ℕ : Type
but is expected to have type
  ℤ → ℕ → ℤ : Type

eric-wieser: Huh, strange; that wasn't what I expected!

eric-wieser (approved):
Thanks!

amit9oct (approved):

PR #206, opened 2024-08-19

ocfnash:


ocfnash:
@GeorgeTsoukalas Should I continue to PR corrections like the ones in this PR and [its predecessor](https://github.com/trishullab/PutnamBench/pull/203) ?

I appreciate these are not easy changes to review. As an open source maintainer myself, I am well aware of the burden of review so I won't mind if you indicate that now is not the time to send such fixes. OTOH if now is the time, I can probably add another 10 this week.

GeorgeTsoukalas:
> Should I continue to PR corrections like the ones in this PR and [its predecessor](https://github.com/trishullab/PutnamBench/pull/203) ?

Please do! As I mentioned in #203, I have been travelling (and arrived very recently) and with the semester starting soon the team is occupied with other things. My plan yesterday was to review this PR today. Feel free to add corrections as you please, we will get to it when able (for the foreseeable future, this should be without significant delay).

lean4/src/putnam_2015_b6.lean
GeorgeTsoukalas, line 1: Interesting, the explainer-line for tsum says "∑' i, f i is the sum of f it exists, or 0 otherwise." Maybe I should raise an issue to modify this? Seems like this is a point of confusion.

lean4/src/putnam_1991_a6.lean
GeorgeTsoukalas, line 1: Don't worry about the review burden, we have collectively done much much more reviewing and a handful of broader changes is no problem. In #193 I went through and modified erroneous usage of `optParam` by introducing the `=` for Props instead of \iff. Would you recommend I go through those and modify = to \iff where applicable?

lean4/src/putnam_2015_b6.lean
ocfnash: Yes, if you open a pull request clarifying then I will merge it, not least because that doc string also contains a typo (missing the word "if").

Perhaps something like: "`∑' i, f i` is the sum of `f` if it exists and is unconditionally convergent, or 0 otherwise."

GeorgeTsoukalas: The PR appears in [#16219](https://github.com/leanprover-community/mathlib4/pull/16219/files).

GeorgeTsoukalas (approved):
All modifications appear to be accurate upon review. Thank you again for your contribution!

GeorgeTsoukalas (approved):

PR #208, opened 2024-08-29

ocfnash:


lean4/src/putnam_2019_a4.lean
```diff
@@ -1,12 +1,12 @@
...
 
 abbrev putnam_2019_a4_solution : Prop := sorry
 -- False
 theorem putnam_2019_a4
-(fint : ((EuclideanSpace ℝ (Fin 3)) → ℝ) → Prop)
-(hfint : ∀ f : (EuclideanSpace ℝ (Fin 3)) → ℝ, fint f = (∀ S : EuclideanSpace ℝ (Fin 3), (∫ x in {p : EuclideanSpace ℝ (Fin 3) | dist p S = 1}, f x) = 0))
-: (∀ f : (EuclideanSpace ℝ (Fin 3)) → ℝ, (Continuous f ∧ fint f) → (∀ x : EuclideanSpace ℝ (Fin 3), f x = 0)) ↔ putnam_2019_a4_solution :=
-sorry
+    (P : (EuclideanSpace ℝ (Fin 3) → ℝ) → Prop)
+    (P_def : ∀ f, P f ↔ ∀ C, integral μH[2] ((sphere C 1).restrict f) = 0) :
```
eric-wieser, line 10: Isn't there a notation for integrating with respect to a measure?

ocfnash: Oh indeed there is, so what I have written is defeq to `∫ x, (sphere C 1).restrict f x ∂μH[2]`. However I now realise that this is equal to `∫ x in sphere C 1, f x ∂μH[2]` (via `MeasureTheory.integral_subtype_comap` and some messing).

lean4/src/putnam_2008_b3.lean
GeorgeTsoukalas, line 1: If C is taken to be arbitrarily far from A, then one can set r arbitrarily large, right? In this case, sphere C r \cap A = \varnothing \subset H. I think one would want to enforce C \in A, otherwise the circle inside H may not be a great circle and hence the radius it actually contributes is less than r.

ocfnash: Great catch: you're absolutely right.

I have fixed this, rewritten the git history, and force pushed.

GeorgeTsoukalas (approved):
All changes look good to me. Thanks for the contribution!

PR #210, opened 2024-09-04

ocfnash:


lean4/src/putnam_1969_a5.lean
```diff
@@ -4,7 +4,16 @@ open BigOperators
...
-sorry
+    (x₀ y₀ t : ℝ)
+    (ht : 0 < t) :
+    x₀ = y₀ ↔ ∃ x y u : ℝ → ℝ,
+      Differentiable ℝ x ∧
+      Differentiable ℝ y ∧
+      Continuous u ∧
+      deriv x = - 2 • y + u ∧
+      deriv y = - 2 • x + u ∧
+      x 0 = x₀ ∧
```
GeorgeTsoukalas, line 15: We tried to avoid using subscript notation `x\0` in the case that certain approaches might not have such tokens in their vocabulary, or are not frequent in their training data. Do you think this is a valid concern?

lean4/src/putnam_1968_a3.lean
```diff
@@ -1,12 +1,12 @@
...
-(S : Finset a)
-(ha : SDiff (Finset a))
-: ∃ l : List (Finset a), (∀ i : Fin l.length, l[i] ⊆ S) ∧ l[0]! = ∅ ∧ (∀ s ⊆ S, ∃! i : Fin l.length, l[i] = s) ∧
-∀ i ∈ Finset.range (l.length - 1), (l[i]! ⊆ l[i+1]! ∧ (l[i+1]! \ l[i]!).card = 1) ∨ (l[i+1]! ⊆ l[i]! ∧ (l[i]! \ l[i+1]!).card = 1) :=
-sorry
+    (α : Type*) [Finite α] :
+    ∃ (n : ℕ) (s : Fin (2 ^ n) → Set α),
+      s 0 = ∅ ∧
+      (∀ t, ∃! i, s i = t) ∧
+      (∀ i, i + 1 < n → (s i ∆ s (i + 1)).ncard = 1) := by
```
GeorgeTsoukalas, line 11: Think this should be `i+1 < 2^n`, right?

lean4/src/putnam_1963_a4.lean
```diff
@@ -1,12 +1,12 @@
...
-(apos : (ℕ → ℝ) → Prop)
-(hapos : apos = fun a => ∀ n, a n > 0)
-(f : (ℕ → ℝ) → ℕ → ℝ)
-(hf : f = fun (a : ℕ → ℝ) n => n * (((1 + a (n+1)) / (a n)) - 1))
-: (∀ a, apos a → limsup (f a) atTop ≥ 1) ∧ (¬∃ c > 1, ∀ a, apos a → limsup (f a) atTop ≥ c) :=
-sorry
+    (T : (ℕ → ℝ) → (ℕ → ℝ))
+    (T_def : ∀ a n, T a n = n * ((1 + a (n + 1)) / a n - 1))
+    (P : (ℕ → ℝ) → ℝ → Prop)
+    (P_def : ∀ a C, P a C ↔ C ≤ limsup (T a) atTop ∨ ¬ BddAbove (range a)):
```
GeorgeTsoukalas, line 10: Shouldn't this be `¬ BddAbove (range (T a))`? Roughly, if `a` is a constant sequence, then `T a` is unbounded.

lean4/src/putnam_1969_a5.lean
ocfnash: We use unicode very heavily through Mathlib so I think this is not a concern. However for the sake of consistency with the rest of PutnamBench, I have removed these subscripts.

lean4/src/putnam_1968_a3.lean
ocfnash: Yes, yikes. Thanks!

GeorgeTsoukalas (approved):
All proposed modifications look good to me. Beyond corrections to misformalizations, the new modifications are also easy on the eyes. Thanks for the contribution!

PR #212, opened 2024-09-13

ocfnash:


lean4/src/putnam_1984_a6.lean
```diff
@@ -1,18 +1,18 @@
...
-(hgpeq : gpeq = fun g : ℕ → ℕ => fun p : ℕ => (p > 0 ∧ ∀ s ≥ 1, g s = g (s + p)))
-(hf : ∀ n > 0, f n = if (n % 10) ≠ 0 then n % 10 else f (n / 10))
-: ∃ g : ℕ → ℕ, (∀ (k : ℕ) (a : ℕ → ℕ), kadistinct k a → g (∑ i : Fin k, a i) = f (∑ i : Fin k, 5 ^ (a i))) ∧
-(if putnam_1984_a6_solution.1 = True then (gpeq g putnam_1984_a6_solution.2 ∧ ∀ p : ℕ, gpeq g p → p ≥ putnam_1984_a6_solution.2) else (¬∃ p : ℕ, gpeq g p)) :=
-sorry
+    (f : ℕ → ℕ)
+    (hf : ∀ n, some (f n) = (Nat.digits 10 (n !)).find? (fun d ↦ d ≠ 0))
+    (P : (ℕ → ℕ) → ℕ → Prop)
+    (P_def : ∀ g p, P g p ↔
+      if p = 0 then (∀ q > 0, ¬ Periodic g q) else (Periodic g p ∧ ∀ q > 0, Periodic g q → p ≤ q)) :
```
GeorgeTsoukalas, line 14: Might this not be a case to use `IsLeast`?

lean4/src/putnam_1963_a3.lean
```diff
@@ -1,21 +1,17 @@
...
-(hP : P 0 y = y ∧ ∀ m ∈ Finset.range n, P (m + 1) y = D (n - 1 - m) (P m y))
-: (ContDiff ℝ n y) ∧ (∀ x ≥ 1, P n y x = f x) ∧ (∀ i ∈ Finset.range n, iteratedDeriv i y 1 = 0) :=
-sorry
+    (P : ℕ → (ℝ → ℝ) → (ℝ → ℝ))
+    (hP : P 0 = id ∧ ∀ i y, P (i + 1) y = P i (fun x ↦ x * deriv y x - i * y x))
+    (n : ℕ)
+    (hn : 0 < n)
+    (f y : ℝ → ℝ)
+    (hf : ContinuousOn f (Ici 1)) :
+    ContDiffOn ℝ n y (Ici 1) ∧ (∀ i < n, deriv^[i] y 1 = 0) ∧ P n y = f ↔
```
GeorgeTsoukalas, line 15: Should it not be that P n y and f are equal on [1, \infty)? I am not sure it matters either way but may be closer to the informal statement.

lean4/src/putnam_1984_a6.lean
ocfnash: Good idea; done.

lean4/src/putnam_1963_a3.lean
ocfnash: As you rightly point out, technically it doesn't matter. However I tend to agree so I have made this change.

ocfnash: Actually I now realise your suggestion is necessary for the converse direction to be true: good catch!

GeorgeTsoukalas (approved):
Proposed modifications look good to me. I have detected no errors, so hopefully that means w.h.p. there are none. Thanks for the contribution!

PR #214, opened 2024-09-21
Move to mathcomp (69-72) & Lean slight modifications

GeorgeTsoukalas:


eric-wieser (approved):
The lean changes look harmless to me, I can't comment on the Coq ones

GeorgeTsoukalas:
Thanks for giving a look over. The Coq ones are as good as they can be without having feedback from a seasoned mathcomp user. Unfortunately, nobody has given any feedback on the mathcomp formalizations. I will probably leave this up for a couple days to allow for that, and then merge it and dump the next set of mathcomp formalizations along with other various changes.

Issue #196, opened 2024-08-06
Feature request: include the original problem descriptions in docstrings for the theorem statements


eric-wieser:
It's quite hard to spot misformalizations when reading the formalization alone.
It would be great if the $$LaTeX$$ description could be included in a docstring (`/-- doc -/` in lean, I can't speak for other systems) for the theorems, to make mistakes easier to spot and correct.

LasseBlaauwbroek:
Seconded for Coq: `(** latex *)`.

eric-wieser:
Probably also a good idea to have a script that extracts the docstrings from each language, and checks in CI that they are all consistent.

amit9oct:
Thank you for opening this issue. I agree that something like this will make it easy for the repository maintainers to check the correctness of the formalization itself. 

We are planning to add a tool (probably a web page on our main website) that will grab the three formalizations along with informal statements from the GitHub repository and then show them side-by-side. We can also build a local version of this so that the maintainers can see these on their local machines.

The reason for not adding comments in the file is that it will make it hard to maintain the same informal statement in two places, and might create problems when one gets modified but the other does not.  If we create a tool for this, then we will not have to deal with any consistency issues which may arise. This will also keep parsing of formalizations simple, for example, if someone wants to just use the formal statement and not use informal statements for their AI tests they will not have to parse and remove the informal statement from the file.

eric-wieser:
> The reason for not adding comments in the file is that it will make it hard to maintain the same informal statement in two places, and might create problems when one gets modified but the other does not.

My recommendation here would be:
* Write a script for each of Coq, Lean, and Isabelle that extracts the doc comment to a json file, mapping problem names to docstrings. I suggest separate scripts since you might find that it is best to parse the Coq code with Coq, the Lean code with Lean, etc.
* Add a test in github's CI that verifies the outputs of these three scripts are identical

CI is a fantastic tool for enforcing consistency.

eric-wieser:
> CI is a fantastic tool for enforcing consistency.

#216 adds the docstrings to all the lean files, and uses CI to ensure they are the same as the ones in the json file.

PR #216, opened 2024-09-23
Add the docstrings to the Lean files.

eric-wieser:
Also includes a CI step to verify the docstrings are right.

If docstrings are missing, this script adds them. If docstrings do not match the JSON, this script emits a github annotation on the relevant file in the PR.

All the changes to the `.lean` files in this PR are from this script.

Addresses #196 for Lean only.


GeorgeTsoukalas:
My feeling is that perhaps it would be better to have the docstrings appear before the first "real" content in each file. This would be after the `open Mathlib` and whatever else is opened, but before the solution if it exists. Also, having the docstring appear between a note about a formalization and the formalization might be somewhat discontinuous. What do you think? I'm not familiar with the convention in Mathlib.

eric-wieser:
> but before the solution if it exists.

Arguably this is already a strang

> but before the solution if it exists.

I agree that having the informal question after the solution is a bit weird, but I'd also contend that it is weird to have the formal question after the formal solution!

A way around this is to write:
```lean
def Putnam_1962_A5 (solution : ℕ → ℕ) : Prop :=
  ∀ n ≥ 2, solution n = ∑ k in Finset.Icc 1 n, Nat.choose n k * k^2 :=

-- easy mode
theorem putnam_1962_a5 : Putnam1962A5 fun n : ℕ => n * (n + 1) * 2^(n - 2) := sorry

-- hard mode
theorem putnam_1962_a5 : Putnam1962A5 sorry := sorry
```
which surely works in Coq and probably in Isabelle as well.

eric-wieser:
(One advantage of such a scheme is it allows stating the predicate "this statement has a unique solution")

amit9oct:
Thank you for adding the PR. I have a suggestion, can we add a warning in the comment that it is auto-generated and should not be changed manually? This will help new users prevent breaking the CI and then realizing that they were not supposed to change this. For example

```lean
/--
WARNING START: AUTO-GENERATED comment. PLEASE DON'T CHANGE.
Find every real-valued function $f$ whose domain is an interval $I$ (finite or infinite) having 0 as a left-hand endpoint, such that for every positive member $x$ of $I$ the average of $f$ over the closed interval $[0, x]$ is equal to the geometric mean of the numbers $f(0)$ and $f(x)$.
WARNING END
-/
theorem putnam_1962_a2
    (P : Set ℝ → (ℝ → ℝ) → Prop)
    (P_def : ∀ s f, P s f ↔ 0 ≤ f ∧ ∀ x ∈ s, ⨍ t in Ico 0 x, f t = √(f 0 * f x)) :
```
This will also help users parse out the comment programmatically (if they need to in their testing) based on the start and end message of this particular comment)

eric-wieser:
> This will help new users prevent breaking the CI and then realizing that they were not supposed to change this. For example

How about I just make the CI give a clearer message? "PLEASE DON'T CHANGE." is the wrong message; it's fine to change it, you just have to change `informal.json` at the same time.

eric-wieser:
I'm testing this out now in https://github.com/eric-wieser/PutnamBench/pull/1/files; the result should be an annotation on the theorem with the bad docstring.

amit9oct:
> > This will help new users prevent breaking the CI and then realizing that they were not supposed to change this. For example
> 
> How about I just make the CI give a clearer message? "PLEASE DON'T CHANGE." is not really the right message; it's fine to change it, you just have to change `informal.json` at the same time.

I still believe any auto-generated code/comment must mention that it is auto-generated. You can change the message, saying that one shouldn't directly modify it here, and refer to some notes in the README about policy for adding or changing informal statements. Even if you put a clear message in CI, this will only make the user realize the mistake after the CI fails, and merge is not allowed.

eric-wieser:
> I still believe any auto-generated code/comment must mention that it is auto-generated. 

I would claim this is not auto-generated going forward; I have initialized them with a script, but they are now auto-synchronized.

> Even if you put a clear message in CI, this will only make the user realize their mistake after the CI fails, and merge is not allowed.

I'm a little confused by this claim; this is the whole point in CI! If the user makes a typo in their Lean code, they may also only realize their mistake after the CI fails.

>  some notes in the README about policy for adding or changing informal statements.

This is a good idea, I will do this tomorrow.

amit9oct:
> I would claim this is not auto-generated going forward; I have initialized them with a script, but they are now just checked for synchronization.
> 
Everyone should then probably use this script going forward, why manually make changes at two places, creating a possibility of error only later to be caught during the CI?
 
> I'm a little confused by this claim; this is the whole point in CI! If the user makes a typo in their Lean code, they may also only realize their mistake after the CI fails.
Breaking of CI helps us ensure that nothing bad gets checked in, but for the users to realize that they don't need to change the comment they will have to wait to create a PR and the workflow to finish running.  Let's say after changing the comment on 10 files and raising a PR, someone realizes their mistake after seeing the CI message, then they have to change all the 10 files again, push a new PR, and ensure that CI is working this time. All this can be avoided by just a message to read the README before changing the comment. This will save a lot of time for any contributor.  This also helps set the expectations right for the user, if we want them to follow something there should be a clear message about it.

Basically, one should not wait for CI to fail to realize mistakes that can be easily avoided and save time on user end.

amit9oct:
>I don't understand this claim. If a PR fails CI, you don't throw away the PR any make a new one; you just push another commit to the old one fixing the error. In this situation, that would mean:

My bad one needs to push another commit, still it is extra wait time to know that CI failed and then wait for another CI to run.

>* You did not realize you made this change:
>  * You use `git checkout -p` to revert just your docstring changes, keeping the rest.

This won't be as simple as `git checkout` because there might be other changes along with the comment the user would want to preserve. I think _just a message to read the README on every comment should be enough_. It improves readability and ensures that it does not frustrate the user to redo the changes again because they missed this extra check we do in our repo. It is not natural for Lean comments to be tested against something, it sets the expectations correctly for the user.

amit9oct:
> > This won't be as simple as git checkout because there might be other changes along with the comment the user would want to preserve.
> 
> This is what the `-p` flag does; it lets you add and remove partial changes.

The partial changes have to pickuped and a new commit has to pushed and one needs to **wait for another CI run**. Maybe I'm not able to get my thoughts spelled out correctly. It is not natural for comments to be parsed and checked. So we must warn the user about it. Most people assume natural things about code like compilation and execution of testcases. Even for those who read the whole README carefully, they might forget to remember this detail and then wait for the CI to fail and then again wait for another CI (no matter how they decide to fix their mistake). It reduces the barrier to contribution, and ensures smooth experience.

eric-wieser:
@amit9oct, I think you're worried about a case that is almost never going to happen generating extra work.
I would be very surprised if anyone contributing to this repository changes the descriptions accidentally; they know that these are supposed to match the pristine putnam questions, and so cannot just decide to change the wording for stylistic reasons.

I think the two categories of contributor that you will have are:
* People who spot escaping errors in the informal text (#218). I think this is going to be very rare now that they have already been fixed.
* People fixing a misformalization like @ocfnash . These people benefit hugely from being able to see the docstring in the same file as the statement, and would quickly get annoyed by a reminder to read the readme in every single docstring.

Yes, it is annoying when you have to wait for a CI run to see if something worked, but we're talking about ~5 minutes of waiting from a small minority of contributors.

amit9oct:
> @amit9oct, I think you're worried about a case that is almost never going to happen generating extra work. I would be very surprised if anyone contributing to this repository changes the descriptions accidentally; they know that these are supposed to match the pristine putnam questions, and so cannot just decide to change the wording for stylistic reasons.
> 
> I think the two categories of contributor that you will have are:
> 
> * People who spot escaping errors in the informal text ([Fix some escaping issues in informal LaTeX. #218](https://github.com/trishullab/PutnamBench/pull/218)). I think this is going to be very rare now that they have already been fixed.
> * People fixing a misformalization like @ocfnash . These people benefit hugely from being able to see the docstring in the same file as the statement, and would quickly get annoyed by a reminder to read the readme in every single docstring.
> 
> Yes, it is annoying when you have to wait for a CI run to see if something worked, but we're talking about ~5 minutes of waiting from a small minority of contributors.

Although I don't completely agree with your idea. However, given only a small number of people are contributing now so I think it is okay. However, we should a have a warning note on README (preferably highlighted), so that any new contributor knows about it.

eric-wieser:
>  However, we should a have a warning note on README

I already added a note to `lean4/README.md` and `informal/README.md`; could you check if they look ok?

amit9oct (commented):
We can say that failing to do so will lead to CI fail and PR won't get merged unless it is fixed.

lean4/README.md
```diff
@@ -7,7 +7,11 @@ import Mathlib
 abbrev putnam_easy_solution : ℕ := sorry
 -- 2
 
+/-- What is 1 + 1 -/
 theorem putnam_easy : 1 + 1 = putnam_easy_solution := sorry
 ```
 
 To generate versions with the solution inlined into the question, you can use `python scripts/rewrite_solutions.py`.
+
+The docstring on the theorem must stay in sync with `informal/putnam.json`,
```
amit9oct, line 16: We can mention that it is checked by CI and out-of-sync comments will not let the PR get merged.

eric-wieser: I've added a sentence about this being checking in CI.

I have not added anything saying "CI has to pass for PRs to be merged" because this is like writing "You should read the README" or similar - it's implied in the act of contributing to anything on github.

If you really feel strongly about this mention, you can tick this box in the repo settings:

image


which will make the github PR page say "this can't be merged because CI fails" when CI fails.

amit9oct: It is okay. I wanted to make sure that people at least know this is checked, which was not mentioned earlier.

amit9oct (approved):

GeorgeTsoukalas:
Having read through the conversation, I tend to agree with the assessment that a mention in the README is sufficient to indicate that the docstring must stay in sync with informal.json. It is probably unlikely that there will be contributions from a large group of people not-already-contributing, so this should be sufficient. We can always change it if necessary, but my feeling is that having extra boilerplate text in each formalization is unnecessary at this point.

> I agree that having the informal question after the solution is a bit weird, but I'd also contend that it is weird to have the formal question after the formal solution!

This is a good point, though I am not sure if at this stage it is compelling enough reason to modify the structure of each problem. We could discuss this further, but maybe that could be in a different PR. It might be a fair amount of work to make similar changes to Isabelle + Coq. 

> Is there an example in the diff? Happy to change this, it doesn't affect the docstring attachment.

I think this might affect the docstring attachment, and the difference is quite minor, so I think it is fine as is.

eric-wieser:
> This is a good point, though I am not sure if at this stage it is compelling enough reason to modify the structure of each problem. We could discuss this further, but maybe that could be in a different PR. It might be a fair amount of work to make similar changes to Isabelle + Coq.

I completely agree this is out of scope for this PR, and this is not the right place to discuss it!

PR #217, opened 2024-09-24
Fix one Lean misformalisation

ocfnash:


ocfnash:
I've just realised that I presumably have to add the informal statements for the four new problems also. I'll come back to this tomorrow.

GeorgeTsoukalas:
To clarify, these are "missing" in the sense that we never formalized them, right? I don't believe we've formalized these in any of the other languages either. Presumably there was a barrier in our understanding or in the naturality of the formalization.

lean4/src/putnam_2023_a4.lean
```diff
@@ -0,0 +1,15 @@
...
+
+theorem putnam_2023_a4
+    (v : Fin 12 → EuclideanSpace ℝ (Fin 3))
+    (hv :
+      letI α : ℝ := (1 + √5) / 4
+      letI h : ℝ := 1 / 2
+      ∃ g : EuclideanSpace ℝ (Fin 3) ≃ₗᵢ[ℝ] EuclideanSpace ℝ (Fin 3),
+      g ∘ v = ![![h, α, 0], ![-h,  α,  0], ![ h, -α,  0], ![-h, -α,  0],
+                ![α, 0, h], ![ α,  0, -h], ![-α,  0,  h], ![-α,  0, -h],
+                ![0, h, α], ![ 0, -h,  α], ![ 0,  h, -α], ![ 0, -h, -α]])
```
eric-wieser, line 11: Strictly spealing these are ill-typed; each of the inner `![]`s needs a `PiLp.equiv`

lean4/src/putnam_2023_b3.lean
```diff
@@ -0,0 +1,18 @@
+import Mathlib
+
+open ProbabilityTheory Set
+
+noncomputable abbrev putnam_2023_b3_solution : ℕ → ℝ := sorry
+-- (fun n : ℕ ↦ (2 * (n : ℝ) + 2) / 3)
+
+theorem putnam_2023_b3
+    (IsZigZag : {k : ℕ} → (Fin k → ℝ) → Prop)
+    (IsZigZag_def : ∀ (k : ℕ) (hk : NeZero k) (y : Fin k → ℝ),
```
eric-wieser, line 10: Maybe
```suggestion
    (IsZigZag_def : ∀ (k : ℕ) [NeZero k] (y : Fin k → ℝ),
```

lean4/src/putnam_2023_a4.lean
ocfnash: Fair enough, I've adjusted and (force) pushed.

lean4/src/putnam_2023_b3.lean
ocfnash: I remember considering this but I can't remember my reservations so I've adjusted and (force) pushed.

ocfnash:
> To clarify, these are "missing" in the sense that we never formalized them, right?

Yes, exactly.

> I don't believe we've formalized these in any of the other languages either.

Agreed.

> Presumably there was a barrier in our understanding or in the naturality of the formalization.

I can't comment here but they are definitely slightly awkward to state.

GeorgeTsoukalas:
Thanks for taking this on! Do you think you will be adding more new formalizations? I might need to add other team members to assist with reviewing, and also possibly to add Isabelle formalizations, if this is the case.

eric-wieser:
Perhaps in that case the fixed formalization should be split to a separate PR from the new ones?

ocfnash:
> Perhaps in that case the fixed formalization should be split to a separate PR from the new ones?

Agreed, done. The new statements now exists in [this branch](https://github.com/ocfnash/PutnamBench/tree/add_missing_statements) and I will create a PR if / when George indicates he is open to this (as well as after I have had a chance to add the informal statements).

ocfnash:
> Do you think you will be adding more new formalizations?

I'm not sure, for now I have other priorities but things may change. Sorry not to have a better answer.

> I might need to add other team members to assist with reviewing, and also possibly to add corresponding Isabelle formalizations, if this is the case.

Totally understandable. For now I'll assume you are not open to adding new statements but I might ask you again in the near future to see if that changes (no problem if not!).

ocfnash:
For emphasis: this PR now has just one commit with a small fix.

GeorgeTsoukalas:
> Totally understandable. For now I'll assume you are not open to adding new statements but I might ask you again in the near future to see if that changes (no problem if not!).

By default I am definitely open to contributions to the benchmark! Please feel free to contribute new statements as you please, I'll handle on our end coordinating who on the team can assist with reviews in the case that effort required is such that I alone cannot make the review process speedy and effective while balancing other priorities.

GeorgeTsoukalas (approved):

PR #218, opened 2024-09-25
Fix some escaping issues in informal LaTeX.

elkhrt:
Backslashes need escaping in JSON strings - fixing several occurrences of this.


eric-wieser:
This probably impacted #216; thanks for catching this!

GeorgeTsoukalas (approved):
Must have missed these when we initially put this together. Thanks for catching these and the contribution!

PR #220, opened 2024-09-26
Add a readme for the lean4 folder

eric-wieser:
I've split this from #216, since it is not dependent on the changes there.

@amit9oct, are you happy with this?

### DIFF
--- a/lean4/README.md
+++ b/lean4/README.md
@@ -1,0 +1,13 @@
+# Lean formalizations
+
+The formalizations in this folder are of the form
+```lean
+import Mathlib
+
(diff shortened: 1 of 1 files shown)


amit9oct:
We should also mention that CI won't succeed if the problem statement in comments in `.lean` file matches the description in the `json`

eric-wieser:
This is not true in master! I'm suggesting this readme be merged first, and I add the sentence you describe in the other PR :)

amit9oct (approved):

amit9oct:
Approved

PR #221, opened 2024-09-26
Warn about LaTeX escaping in json

eric-wieser:
This is a pretty easy mistake to make (c.f. #218), so perhaps worth documenting.

I'll leave it to @amit9oct to decide.


amit9oct:
Thank you for documenting it. Approved.

amit9oct (approved):

PR #222, opened 2024-10-02
Update to Lean 4.11.0

ocfnash:
The changes to the five files:
 * `putnam_1965_a5.lean`
 * `putnam_1985_b1.lean`
 * `putnam_1994_a6.lean`
 * `putnam_2011_b6.lean`
 * `putnam_2016_b2.lean`

are fixes to account for the fact that Lean now tries to interpret certain expressions of the form `{ ... }.ncard` using `Finset.ncard` (which does not exist) rather than `Set.ncard`. Similar remarks for `Set.encard`. I have changed these to use `Finset.card`.


lean4/src/putnam_1985_b1.lean
```diff
@@ -13,9 +13,11 @@ p(x) = (x-m_1)(x-m_2)(x-m_3)(x-m_4)(x-m_5)
...
-(hp : p = fun m ↦ ∏ i : Fin 5, ((X : Polynomial ℝ) - m i))
-(numnzcoeff : Polynomial ℝ → ℕ)
-(hnumnzcoeff : numnzcoeff = fun p ↦ {j ∈ Finset.range (p.natDegree + 1) | coeff p j ≠ 0}.ncard)
-: (Injective putnam_1985_b1_solution ∧ ∀ m : Fin 5 → ℤ, Injective m → numnzcoeff (p putnam_1985_b1_solution) ≤ numnzcoeff (p m)) :=
-sorry
+    (p : (Fin 5 → ℤ) → (Polynomial ℝ))
+    (hp : ∀ m, p m = ∏ i, ((X : Polynomial ℝ) - m i))
+    (numnzcoeff : Polynomial ℝ → ℕ)
+    (hnumnzcoeff : ∀ p, numnzcoeff p =
+      Set.ncard {j ∈ Finset.range (p.natDegree + 1) | coeff p j ≠ 0}) :
```
eric-wieser, line 20: I guess these could switch to using `.card` if the set comprehension is now a `Finset`?

ocfnash: Yes, I was trying to be ultraconservative and not change the meaning but if George prefers I could make such changes.

GeorgeTsoukalas: Yeah I agree it makes sense to switch to .card in this setting. We can also make those changes if it's easier.

ocfnash: OK I've changed these to use `Finset.card`.

GeorgeTsoukalas:
> Lean now tries to interpret certain expressions of the form { ... }.ncard using Finset.ncard (which does not exist) rather than Set.ncard

Could you point me to the relevant changes in the v4.11 release notes that cause this? I want to better understand the cases when this might occur.

eric-wieser:
These won't be in the Lean v4.11.0 release notes, as they are due to a change in Mathlib. Mathlib has no release notes, though I can point to https://github.com/leanprover-community/mathlib4/pull/11582 as the likely cause.

GeorgeTsoukalas (approved):
Looks good, thanks for the contribution!

PR #223, opened 2024-10-04
More minor fixes to informal problem solutions.

elkhrt:


GeorgeTsoukalas (changes requested):
Thanks for pointing out these issues! In general we've tried to keep the format to be "Show that a/the solution is _" so that this could be easily used for someone trying to do informal proof generation conditioned on the answer. Might you be willing to modify 2015_a2 and 2019_b6 to generally reflect this format? I can also specify this in the README

elkhrt:
FWIW, it would be helpful for me to have a simple list of possible answers for the problems which are in the form of a question. I can easily do that transformation locally of course.

GeorgeTsoukalas:
Here is a set of examples (so long as `informal_solution` contains an imperative sentence which is reasonable in context I think the exact wording can be up to interpretation):

1. Problem does not require exhibiting a solution; Convention: "None."
2. Find all x such that P x; Convention: "Show that the set of solutions is _"/"Show that the statement is true for all x in _"
3. Find an x such that P x; Convention: if the Lean formalization exhibits only a single x (though multiple were possible), for better alignment then "Show that a solution is _" else "Show solutions can be taken from _"
4. Is the statement P true? Convention; "Show that the statement is {true/false}" or "Show that 'P is true'"
5. Find the minimum value of x such that P x; Convention: "Show that the minimum value is _"

Again this is mostly up to interpretation and based on the exact stating of the question. The only example I might amend would be "Every n \geq 1" -> "Show that the statement is true for all n \geq 1" to fit an imperative form.

It is a bit regrettable that we let some issues slip through on the informal end, thanks for looking for them! I'll also do a pass myself to see if there's any further declarative informal_solutions.

elkhrt:
Done.

GeorgeTsoukalas:
Looks like the check failed. Think there's an escaping issue in the latest "Rephrasing" commit

eric-wieser:
> 1. Problem does not require exhibiting a solution; Convention: "None."

Maybe using the JSON `null` would be more appropriate for these, to avoid ambiguity with "how many such functions exist? answer: none"

informal/putnam.json
```diff
@@ -4880,7 +4880,7 @@
     {
         "problem_name": "putnam_2019_b6",
         "informal_statement": "Let \\( \\mathbb{Z}^n \\) be the integer lattice in \\( \\mathbb{R}^n \\). Two points in \\( \\mathbb{Z}^n \\) are called neighbors if they differ by exactly 1 in one coordinate and are equal in all other coordinates. For which integers \\( n \\geq 1 \\) does there exist a set of points \\( S \\subset \\mathbb{Z}^n \\) satisfying the following two conditions? \\begin{enumerate} \\item If \\( p \\) is in \\( S \\), then none of the neighbors of \\( p \\) is in \\( S \\). \\item If \\( p \\in \\mathbb{Z}^n \\) is not in \\( S \\), then exactly one of the neighbors of \\( p \\) is in \\( S \\). \\end{enumerate}",
-        "informal_solution": "Show that the above statement is True.",
+        "informal_solution": "Show that the statement is true for every n \geq 1",
```
eric-wieser, line 4883:
```suggestion
        "informal_solution": "Show that the statement is true for every \\(n \\geq 1\\)",
```

elkhrt:
> > 1. Problem does not require exhibiting a solution; Convention: "None."
> 
> Maybe using the JSON `null` would be more appropriate than `"None."` for these, to avoid ambiguity with "how many such functions exist? answer: none"

Or just omitting it entirely.

GeorgeTsoukalas:
Ideally the imperative-style of `informal_solution` when it exists should preclude confusion. Anyways, I think omitting it entirely seems reasonable, writing "None." was a very early design choice in this project. I'll merge this and the omission can be the subject of a different PR. Thank you for the contribution, Edward!

PR #224, opened 2024-10-07
Update to v4.12.0

eric-wieser:
It seems everything works without modification!


GeorgeTsoukalas (approved):
Thanks for the contribution, Eric!

PR #225, opened 2024-10-07
Modify Coq problems to mathcomp & some Lean formalizations.

GeorgeTsoukalas:


lean4/src/putnam_2022_b1.lean
```diff
@@ -6,12 +6,12 @@ open Polynomial
 theorem putnam_2022_b1
 (n : ℕ)
 (P : Polynomial ℝ)
-(B : Polynomial ℝ)
+(B : ℕ → ℝ)
```
eric-wieser, line 9: nit:
```suggestion
(b : ℕ → ℝ)
```
to match the english. I think the uppercase `B` was chosen to be the polynomial with coefficients `b`, but of course that turned otu to be silly.

GeorgeTsoukalas: Removed this from this PR. Can approve and merge #227.

PR #226, opened 2024-10-07
Add missing informal + Lean statements for four missing problems

ocfnash:


lean4/src/putnam_2022_a5.lean
```diff
@@ -0,0 +1,22 @@
...
+      (x = y ∧ ∀ i < 2021, i ∉ x → i + 1 ∈ x) ∨
+      ∃ i < 2021, i ∉ x ∧ i + 1 ∉ x ∧ y = x ∪ {i, i + 1})
+    (IsValidGame : List (Set (Fin 2022)) → Prop)
+    (IsValidGame_def : ∀ g, IsValidGame g ↔ (∃ gt, g = ∅ :: gt) ∧ g.Chain' IsValidMove)
+    (ConformsToStrategy : List (Set (Fin 2022)) → (Set (Fin 2022) → Set (Fin 2022)) → Prop)
+    (ConformsToStrategy_def : ∀ g s, ConformsToStrategy g s ↔
+      ∀ (i) (h : i + 1 < g.length), Even i → g[i + 1] = s g[i]) :
+    IsGreatest
+      {n | ∃ s, (∀ x, IsValidMove x (s x)) ∧ ∀ g,
+        IsValidGame g → ConformsToStrategy g s → ∃ gh x, g = gh ++ [x] → n ≤ xᶜ.ncard}
```
GeorgeTsoukalas, line 20: I think this becomes trivially true for all n, because `∃ gh x, g = gh ++ [x] → n ≤ xᶜ.ncard` should be `∃ gh x, g = gh ++ [x] ∧ n ≤ xᶜ.ncard`, right?

lean4/src/putnam_2023_a6.lean
```diff
@@ -0,0 +1,22 @@
...
+abbrev putnam_2023_a6_solution : Set ℕ := sorry
+-- {n : ℕ | 0 < n}
+/--
+Alice and Bob play a game in which they take turns choosing integers from $1$ to $n$. Before any integers are chosen, Bob selects a goal of 'odd' or 'even'. On the first turn, Alice chooses one of the $n$ integers. On the second turn, Bob chooses one of the remaining integers. They continue alternately choosing one of the integers that has not yet been chosen, until the $n$th turn, which is forced and ends the game. Bob wins if the parity of $\{k : \mbox{the number $k$ was chosen on the $k$th turn}\}$ matches his goal. For which values of $n$ does Bob have a winning strategy?
+-/
+theorem putnam_2023_a6
+    (IsValidGame : List ℕ → Prop)
+    (IsValidGame_def : ∀ g, IsValidGame g ↔ g.Nodup ∧ (∀ i ∈ g, i ∈ Icc 1 g.length))
+    (parityOf : List ℕ → ZMod 2)
+    (parityOf_def : ∀ g, parityOf g = ((range g.length).filter fun i ↦ g[i]! = i + 1).card)
```
GeorgeTsoukalas, line 14: Q: Is it a standard convention to use coercions implicitly from a "larger" type to a "smaller" type? Usually we have used coercions for the other direction. I am curious if this is normal in Mathlib - this might hurt readability slightly.

lean4/src/putnam_2023_a6.lean
```diff
@@ -0,0 +1,22 @@
...
+    (ConformsToStrategy : List ℕ → (List ℕ → ℕ) → Prop)
+    (ConformsToStrategy_def : ∀ g s, ConformsToStrategy g s ↔
+      ∀ (i) (h : i < g.length), Odd i → g[i] = s (g.take i))
+    (IsWinningFor : ℕ → (List ℕ → ℕ) → Prop)
+    (IsWinningFor_def : ∀ n s, IsWinningFor n s ↔
+      ∃ p, ∀ g, g.length = n → IsValidGame g → ConformsToStrategy g s → parityOf g = p) :
```
GeorgeTsoukalas, line 20: While I believe this formalization doesn't have an issue here, I am curious about your thoughts on this question: when writing a formal spec without a matching proof like we do here, should we include "safety blanket" hypotheses? This would give some extra confidence but introduce some clutter since it could be unnecessary. In this case, an example might be the extra constraint on the strategy's behavior (like `s L \notin L`), though I think this follows already.

lean4/src/putnam_2023_b3.lean
```diff
@@ -0,0 +1,20 @@
...
+theorem putnam_2023_b3
+    (IsZigZag : {k : ℕ} → (Fin k → ℝ) → Prop)
+    (IsZigZag_def : ∀ (k : ℕ) [NeZero k] (y : Fin k → ℝ),
+      IsZigZag y ↔ k = 1 ∨ ((∀ i, i + 1 < k → y (i + 1) ≠ y i)) ∧
+        (∀ i, i + 2 < k → y (i + 2) < y (i + 1) ↔ y i < y (i + 1)))
+    (n : ℕ)
+    (hn : 2 ≤ n)
+    (a : (Fin n → Icc (0 : ℝ) 1) → ℕ)
+    (ha : ∀ x, IsGreatest {k | ∃ i : Fin k ↪o Fin n, IsZigZag ((↑) ∘ x ∘ i)} (a x)) :
+    𝔼[(↑) ∘ a] = putnam_2023_b3_solution n := by
```
GeorgeTsoukalas, line 19: Clarification Q: The default probability measure on [0,1]^n in Lean is the lebesgue measure, right?

informal/putnam.json
```diff
@@ -5065,6 +5065,14 @@
             "number_theory"
         ]
     },
+    {
+        "problem_name": "putnam_2022_a5",
+        "informal_statement": "Alice and Bob play a game on a board consisting of one row of 2022 consecutive squares. They take turns placing tiles that cover two adjacent squares, with Alice going first. By rule, a tile must not cover a square that is already covered by another tile. The game ends when no tile can be placed according to this rule. Alice's goal is to maximize the number of uncovered squares when the game ends; Bob's goal is to minimize it. What is the greatest number of uncovered squares that Alice can ensure at the end of the game, no matter how Bob plays?",
+        "informal_solution": "290",
```
GeorgeTsoukalas, line 5071: Could you modify these `informal_solution` fields to have an imperative sentence as the value, as opposed to a declarative one. Generally the format would be `Show that the solution is _` or something similar. I will include some general guidelines on this in an upcoming PR.

lean4/src/putnam_2022_a5.lean
ocfnash: Nice catch!

lean4/src/putnam_2023_a6.lean
ocfnash: I'm afraid I don't understand this question, can you point me at an example where you have used a coercion in the "other direction"?

lean4/src/putnam_2023_b3.lean
ocfnash: Yes, this picks up the product measure of the Lebesgue measure. You can investigate this sort of thing as follows:
```lean
variable (n : ℕ)
#synth MeasureTheory.MeasureSpace (Fin n → Icc (0 : ℝ) 1) -- Will reply `MeasureTheory.MeasureSpace.pi`
```
and from there, look at the definition of `MeasureTheory.MeasureSpace.pi`.

ocfnash: Having said that, it's better to look at API than definitions so this is really the right evidence:
```lean
variable (n : ℕ)

open MeasureTheory

example (s : Fin n → Set (Icc (0 : ℝ) 1)) :
    volume (univ.pi s) = ∏ i, volume (s i) :=
  volume_pi_pi s
```

lean4/src/putnam_2023_a6.lean
ocfnash: IMHO, I we should not include such hypotheses because:

- they might mask an error
- they might contain information which is the duty of the problem solver to deduce
- they are redundant

However, I also believe they would probably salvage the occasional misformalisation. So it's a subjective call really, dependent upon what you view as the job of these formalisations.

informal/putnam.json
ocfnash: Done!

lean4/src/putnam_2023_b3.lean
ocfnash: I guess I should also give you evidence that the measure on $[0, 1]$ is the restriction of the Lebesgue measure. Here's a rather ugly proof:
```lean
open MeasureTheory

example (a b : Icc (0 : ℝ) 1) (hab : a ≤ b) :
    (volume (Icc a b)).toReal = (b : ℝ) - (a : ℝ) := by
  have foo : (volume : Measure (Icc (0 : ℝ) 1)) =
    (volume : Measure ℝ).comap ((↑) : (Icc (0 : ℝ) 1) → ℝ) := rfl
  have bar : MeasurableEmbedding ((↑) : (Icc (0 : ℝ) 1) → ℝ) :=
    MeasurableEmbedding.subtype_coe measurableSet_Icc
  rw [foo, Measure.comap_apply _ Subtype.val_injective
    (fun s hs ↦ bar.measurableSet_image.mpr hs) _ measurableSet_Icc]
  simpa
```

lean4/src/putnam_2023_a6.lean
GeorgeTsoukalas: Ah by this I mean a coercion `\N \to \Z`, for example. Here "smaller" and "larger" are a reference to `\N \subset \Z`. This one _seems_ more natural to me, and I might expect a reader to be able to parse this example easier than a coercion "in the other direction." But this is probably negligible, if not an unreasonable take due to inexperience on my part.

lean4/src/putnam_2023_a6.lean
GeorgeTsoukalas: I tend to agree with you. I think in the case with a problem like this one (where the natural language statement isn't ''easily'' formalizable) there is some ambiguity in what might be the asserted conditions we need to put in the spec. For example "Bob chooses one of the **remaining** integers" could directly translate to something like `s L \notin L`. If one agrees on this, then it is probably better that this appears in the formalization (abstractly, I'm not actually saying there should be changes, to be clear), otherwise one needs to infer that the condition follows from `IsValidGame g` and `ConformsToStrategy g s`. Perhaps it is really a matter of how faithful the formalization is to the natural language statement, which I agree is a subjective call.

lean4/src/putnam_2023_b3.lean
GeorgeTsoukalas: Thank you!

lean4/src/putnam_2023_a6.lean
ocfnash: The job of a coercion is to remove the obligation for us to avoid being explicit about maps for which there is only one possible candidate. Or maybe more precisely, the only possible map that the author could have intended without explicitly saying something else.

The map `ℕ → ℤ` which you highlight is certainly of this nature, and I think the map `ℕ → ZMod 2` is too. In any case, Mathlib has already made this call so I suggest we fall in line.

If you care to see the details, the Infoview tells me that the map is provided by the `NatCast` typeclass and so I ran
```lean
#synth NatCast (ZMod 2)
```
This succeeds with output `Semiring.toNatCast` and after chasing through the class hierarchy a bit I find the map arises because `AddMonoidWithOne` extends `NatCast`.

lean4/src/putnam_2023_a4.lean
```diff
@@ -0,0 +1,19 @@
...
+Let $v_1, \ldots, v_{12}$ be unit vectors in $\mathbb{R}^3$ from the origin to the vertices of a regular icosahedron. Show that for every vector $v \in \mathbb{R}^3$ and every $\epsilon > 0$, there exist integers $a_1, \ldots, a_{12}$ such that $\|a_1v_1 + \cdots + a_{12}v_{12} - v\| < ε$.
+-/
+theorem putnam_2023_a4
+    (v : Fin 12 → EuclideanSpace ℝ (Fin 3))
+    (hv :
+      letI α : ℝ := (1 + √5) / 4
+      letI h : ℝ := 1 / 2
+      letI e : (Fin 3 → ℝ) ≃ EuclideanSpace ℝ (Fin 3) := (WithLp.equiv _ _).symm
+      ∃ g : EuclideanSpace ℝ (Fin 3) ≃ₗᵢ[ℝ] EuclideanSpace ℝ (Fin 3),
+      g ∘ v = e ∘ ![![h, α, 0], ![-h,  α,  0], ![ h, -α,  0], ![-h, -α,  0],
```
eric-wieser, line 13: These strictly aren't well-typed; the inner `![`s should be elements of euclidean space, but here they are raw vectors.

leanprover-community/mathlib4#17732 will give a path to fixing this eventually, but in the meantime you should probably add the missing `(WithLp.equiv _ _).symm` before each one.

GeorgeTsoukalas: The vectors given by (\pm h, \pm a, 0) or any cyclic perm are not actually unit here. Per the solution it doesn't matter but it skips a reasoning step to use this form. (Thanks to Michael for spotting this!)

ocfnash: @eric-wieser

Surely the typing is correct. Indeed `e` has type `(Fin 3 → ℝ) ≃ EuclideanSpace ℝ (Fin 3)` and
```lean
  ![![h, α, 0], ![-h,  α,  0], ![ h, -α,  0], ![-h, -α,  0],
    ![α, 0, h], ![ α,  0, -h], ![-α,  0,  h], ![-α,  0, -h],
    ![0, h, α], ![ 0, -h,  α], ![ 0,  h, -α], ![ 0, -h, -α]]
```
has type `Fin 12 → Fin 3 → ℝ` so after composing with `⇑e` we get a map `Fin 12 → EuclideanSpace ℝ (Fin 3)`.

ocfnash: > The vectors given by (\pm h, \pm a, 0) or any cyclic perm are not actually unit here. 

Thanks. Urgh, it looks like I fixed the edge lengths at 1 instead of placing the vertices on the unit sphere. I've (force) pushed a fix for this.

eric-wieser: Ah apologies, I missed `e` being here

GeorgeTsoukalas (approved):
Everything looks good to me. Thanks so much for the contribution, Oliver!

PR #227, opened 2024-10-07
Fix a misformalization of putnam_2022_b1

eric-wieser:
This falsely stated that exp had a finite polynomial expansion.

This also cleans up the spelling of "with integer coefficients".

### DIFF
--- a/lean4/src/putnam_2022_b1.lean
+++ b/lean4/src/putnam_2022_b1.lean
@@ -7,14 +7,13 @@ open Polynomial
 Suppose that $P(x)=a_1x+a_2x^2+\cdots+a_nx^n$ is a polynomial with integer coefficients, with $a_1$ odd. Suppose that $e^{P(x)}=b_0+b_1x+b_2x^2+\dots$ for all $x$. Prove that $b_k$ is nonzero for all $k \geq 0$.
 -/
 theorem putnam_2022_b1
-(n : ℕ)
-(P : Polynomial ℝ)
-(B : Polynomial ℝ)
-(npos : n ≥ 1)
-(Pconst : P.coeff 0 = 0)
-(Pdegree : P.degree = n)
-(Pint : ∀ k : Set.Icc 1 n, P.coeff k = round (P.coeff k))
-(Podd : Odd (round (P.coeff 1)))
-(hB : ∀ x : ℝ, Real.exp (P.eval x) = B.eval x)
-: ∀ k : ℕ, B.coeff k ≠ 0 :=
-sorry
+    (n : ℕ)
+    (P : Polynomial ℤ)
+    (b : ℕ → ℝ)
+    (npos : n ≥ 1)
+    (Pconst : P.coeff 0 = 0)
+    (Pdegree : P.degree = n)
+    (Podd : Odd (P.coeff 1))
+    (hB : ∀ x : ℝ, HasSum (fun i => b i * x ^ i) (Real.exp (aeval x P))) :
+    ∀ k : ℕ, b k ≠ 0 := by
+  sorry


lean4/src/putnam_2022_b1.lean
```diff
@@ -7,14 +7,13 @@ open Polynomial
...
-: ∀ k : ℕ, B.coeff k ≠ 0 :=
-sorry
+    (n : ℕ)
+    (P : Polynomial ℤ)
+    (b : ℕ → ℝ)
+    (npos : n ≥ 1)
+    (Pconst : P.coeff 0 = 0)
+    (Pdegree : P.degree = n)
+    (Podd : Odd (P.coeff 1))
+    (hB : ∀ x : ℝ, HasSum (fun i => b i * x ^ i) (Real.exp (aeval x P))) :
```
eric-wieser, line 17: I think there might be a preferred spelling of this with `PowerSeries` or `FormalMultilinearSeries`, but I couldn't work it out.

GeorgeTsoukalas:
I actually caught this one yesterday! See #225. Which spelling do you think is better?

eric-wieser:
Ah, nice! I guess either spelling is probably fine, though `tsum` is slightly riskier in general since on non-summable series it is zero.

GeorgeTsoukalas:
Good point. I can remove that commit from my PR and merge this one instead, if you'd like. I tend to prefer the changes in here (and the comment you had about 'B' -> 'b' is already applied here).

lean4/src/putnam_2022_b1.lean
eric-wieser: @ocfnash, do you think HasSum or tsum is better here?

ocfnash: I have a mild preference for `HasSum` here but either is fine.

lean4/src/putnam_2022_b1.lean
```diff
@@ -7,14 +7,13 @@ open Polynomial
...
-(Podd : Odd (round (P.coeff 1)))
-(hB : ∀ x : ℝ, Real.exp (P.eval x) = B.eval x)
-: ∀ k : ℕ, B.coeff k ≠ 0 :=
-sorry
+    (n : ℕ)
+    (P : Polynomial ℤ)
+    (b : ℕ → ℝ)
+    (npos : n ≥ 1)
+    (Pconst : P.coeff 0 = 0)
+    (Pdegree : P.degree = n)
```
ocfnash, line 15: You could drop `n` entirely and just write:
```suggestion
    (Pdegree : 0 < P.degree)
```
Even Lean agrees (warning ugly proof):
```lean
example (P : Polynomial ℤ) (h : Odd (P.coeff 1)) :
    0 < P.degree := by
  have h0 : ¬ Odd (0 : ℤ) := by simp [← Int.not_even_iff_odd, not_not]
  have hP : 0 ≤ P.degree := by
    rw [Polynomial.zero_le_degree_iff]
    rintro rfl
    rw [Polynomial.coeff_zero] at h
    contradiction
  suffices P.degree ≠ 0 from lt_of_le_of_ne hP this.symm
  rintro contra
  rw [Polynomial.eq_C_of_degree_eq_zero contra, Polynomial.coeff_C_succ] at h
  contradiction
```

eric-wieser: Yes, I guess the question is whether including `n` makes this a more literal translation.

GeorgeTsoukalas: I think it can be dropped, I would say the informal statement mostly uses it to describe the form of a polynomial which is unnecessary.

eric-wieser: Done.

GeorgeTsoukalas (approved):
Thanks for the contribution, Eric!

PR #229, opened 2024-10-16

ocfnash:


GeorgeTsoukalas:
Nice catch for noticing a mistake. I think the case for the $r_+$ case may still be inaccurate, in particular I think the term may need to be $\dots - |a|r_+$. Here is a graph of an example: https://www.desmos.com/calculator/e90qfxhkr9.

ocfnash:
Sorry I meant to get back to you about this: will do so tomorrow.

ocfnash:
> I think the case for the r + case may still be inaccurate, in particular I think the term may need to be ⋯ − | a | r + . Here is a graph of an example: https://www.desmos.com/calculator/e90qfxhkr9.

In fact I believe all is well, though I nearly went mad trying to understand why your plot made things look wrong. I somehow FINALLY realised that in this desmos site, `log` is base 10 so we need to use `ln`. (There is also one other sign error in that calculator you provided but this was easy to spot.) Here is a new calculator which I believe demonstrates that all is well: https://www.desmos.com/calculator/mykkqq1f2t

ocfnash:
In the meantime, I have added three more problems where we have identified misformalisations to this PR.

As always, thanks for review and maintenence.

GeorgeTsoukalas:
> In fact I believe all is well, though I nearly went mad trying to understand why your plot made things look wrong. I somehow FINALLY realised that in this desmos site, `log` is base 10 so we need to use `ln`. (There is also one other sign error in that calculator you provided but this was easy to spot.) Here is a new calculator which I believe demonstrates that all is well: https://www.desmos.com/calculator/mykkqq1f2t


Whoops! Sorry for causing confusion, I was trying to avoid it with the graph but had the opposite effect. I am now confident in the change.

lean4/src/putnam_2017_a2.lean
```diff
@@ -1,11 +1,14 @@
...
 -/
 theorem putnam_2017_a2
-(Q : ℕ → ℝ → ℝ)
-(hQbase : ∀ x : ℝ, Q 0 x = 1 ∧ Q 1 x = x)
-(hQn : ∀ n ≥ 2, ∀ x : ℝ, Q n x = ((Q (n - 1) x) ^ 2 - 1) / Q (n - 2) x)
-: ∀ n > 0, ∃ P : Polynomial ℝ, (∀ i : ℕ, P.coeff i = round (P.coeff i)) ∧ Q n = P.eval :=
-sorry
+    (Q : ℕ → ℝ → ℝ)
+    (hQbase : ∀ x : ℝ, Q 0 x = 1 ∧ Q 1 x = x)
+    (hQn : ∀ n, ∀ x, Q n x ≠ 0 → Q (n + 2) x = ((Q (n + 1) x) ^ 2 - 1) / Q n x)
```
GeorgeTsoukalas, line 11: Nice catch. I am not sure the statement is true still, though. Does this not allow the values of Q (n+2) x be essentially unconstrained when Q n x = 0? Maybe this is better stated as operations of formal polynomials?

lean4/src/putnam_2009_b4.lean
```diff
@@ -1,16 +1,17 @@
...
-(V : Set (MvPolynomial (Fin 2) ℝ)) [AddCommGroup V] [Module ℝ V]
-(hV : ∀ P : MvPolynomial (Fin 2) ℝ, P ∈ V ↔ balanced P ∧ P.totalDegree ≤ 2009)
-: (Module.rank V = putnam_2009_b4_solution) :=
-sorry
+    (IsBalanced : MvPolynomial (Fin 2) ℝ → Prop)
+    (IsBalanced_def : ∀ P, IsBalanced P ↔ ∀ r > 0,
+      (∫ t in (0 : ℝ)..(2 * π), eval ![r * cos t, r * sin t] P) / (2 * π * r) = 0)
+    (V : Set (MvPolynomial (Fin 2) ℝ))
+    (V_def : ∀ P, P ∈ V ↔ IsBalanced P ∧ P.totalDegree ≤ 2009) :
+    Module.rank V = putnam_2009_b4_solution :=
```
eric-wieser, line 16: I am extremely surprised this compiles; certainly it is total nonsense.

The LHS here is of type `∀ (M : Type*) [Semiring ↑V] [AddCommMonoid M] [inst : Module (↑V) M], Cardinal`, which is talking about all possible ring structures(!) on `V` and all possible modules over such rings.

eric-wieser: I think the correct version ends:
```
    (V : Submodule ℝ (MvPolynomial (Fin 2) ℝ))
    (V_def : ∀ P, P ∈ V ↔ IsBalanced P ∧ P.totalDegree ≤ 2009) :
    Module.rank ℝ V = putnam_2009_b4_solution := sorry
```

eric-wieser: (#240)

ocfnash: Yiiiikes. Fixed and (force) pushed.

lean4/src/putnam_2017_a2.lean
ocfnash: Good point! I've restated it using rational functions to fix this.

GeorgeTsoukalas (approved):
Everything looks good to me, thanks for the contribution!

PR #230, opened 2024-10-17
Fix 1996 A4

faabian:
This was trivially provable because assumption `hS2` applied to `a`, `b`, `a` simplifies to `False`. (Evidently, `(a, b, a) \notin S`, but that does not disallow simplifying `P \iff not P` to `False` in the first place.)

### DIFF
--- a/lean4/src/putnam_1996_a4.lean
+++ b/lean4/src/putnam_1996_a4.lean
@@ -12,7 +12,7 @@ theorem putnam_1996_a4
 (S : Set (A × A × A))
 (hSdistinct : ∀ a b c : A, ⟨a, b, c⟩ ∈ S → a ≠ b ∧ b ≠ c ∧ a ≠ c)
 (hS1 : ∀ a b c : A, ⟨a, b, c⟩ ∈ S ↔ ⟨b, c, a⟩ ∈ S)
-(hS2 : ∀ a b c : A, ⟨a, b, c⟩ ∈ S ↔ ⟨c, b, a⟩ ∉ S)
+(hS2 : ∀ a b c : A, a ≠ c → ⟨a, b, c⟩ ∈ S ↔ ⟨c, b, a⟩ ∉ S)
 (hS3 : ∀ a b c d : A, (⟨a, b, c⟩ ∈ S ∧ ⟨c, d, a⟩ ∈ S) ↔ (⟨b,c,d⟩ ∈ S ∧ ⟨d,a,b⟩ ∈ S))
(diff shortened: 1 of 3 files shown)


GeorgeTsoukalas:
I think you need to wrap the \iff with parentheses in the Lean one, this will get parsed different. A -> (B \iff C) is not the same as (A -> B) \iff C.

hrmacbeth:
I think there may be another minor issue in the formal statement here.  The problem says, "Let $S$ be **the** set of ordered triples $(a, b, c)$ of distinct elements of a finite set $A$", but the Lean says effectively that $S$ is **a** set of ordered triples of distinct elements:
```lean
(S : Set (A × A × A))
(hSdistinct : ∀ a b c : A, ⟨a, b, c⟩ ∈ S → a ≠ b ∧ b ≠ c ∧ a ≠ c)
```
This could be fixed by changing the condition from → to ↔:
```lean
(S : Set (A × A × A))
(hSdistinct : ∀ a b c : A, ⟨a, b, c⟩ ∈ S ↔ a ≠ b ∧ b ≠ c ∧ a ≠ c)
```

GeorgeTsoukalas:
> Edit: looking further, the problem makes no sense if formalized as "the" rather than "a". I think this may be an error in the phrasing of the problem! I see that in at least two variant sources [1](https://prase.cz/kalva/putnam/psoln/psol964.html) [2](https://yufeizhao.com/a34/fa19/ps/algebra.pdf) the problem is paraphrased to give the "a" meaning.

Good catch! I'll make a change to the informal statement because of this poor wording.

lean4/src/putnam_1996_a4.lean
```diff
@@ -12,7 +12,7 @@ theorem putnam_1996_a4
 (S : Set (A × A × A))
 (hSdistinct : ∀ a b c : A, ⟨a, b, c⟩ ∈ S → a ≠ b ∧ b ≠ c ∧ a ≠ c)
 (hS1 : ∀ a b c : A, ⟨a, b, c⟩ ∈ S ↔ ⟨b, c, a⟩ ∈ S)
-(hS2 : ∀ a b c : A, ⟨a, b, c⟩ ∈ S ↔ ⟨c, b, a⟩ ∉ S)
+(hS2 : ∀ a b c : A, a ≠ c → ⟨a, b, c⟩ ∈ S ↔ ⟨c, b, a⟩ ∉ S)
```
GeorgeTsoukalas, line 15:
```suggestion
(hS2 : ∀ a b c : A, a ≠ c → (⟨a, b, c⟩ ∈ S ↔ ⟨c, b, a⟩ ∉ S))
```

GeorgeTsoukalas (approved):
Everything looks good. I will add the informal statement modification to 1996_b4 in an upcoming PR. Thank you for the contribution, Fabian!

PR #233, opened 2024-10-21
Use `⟪x, y⟫_ℝ` instead of `Matrix.dotProduct`

eric-wieser:
Strictly `dotProduct` is for plain functions, not for elements of `EuclideanSpace`.

The original phrasing was still mathematically correct, as shown by [`EuclideanSpace.inner_eq_star_dotProduct`](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Analysis/InnerProductSpace/PiL2.html#EuclideanSpace.inner_eq_star_dotProduct) (and `star x = x` on the real numbers); but it is using "evil" casting between `EuclideanSpace` and `Function`.


lean4/src/putnam_1998_a6.lean
```diff
@@ -13,6 +14,6 @@ theorem putnam_1998_a6
 (hint : ∀ i : Fin 2, ∃ a b c : ℤ, A i = a ∧ B i = b ∧ C i = c)
 (htriangle : A ≠ B ∧ A ≠ C ∧ B ≠ C)
 (harea : (dist A B + dist B C) ^ 2 < 8 * (MeasureTheory.volume (convexHull ℝ {A, B, C})).toReal + 1)
-(threesquare : (EuclideanSpace ℝ (Fin 2)) → (EuclideanSpace ℝ (Fin 2)) → (EuclideanSpace ℝ (Fin 2)) → Prop := fun P Q R ↦ dist Q P = dist Q R ∧ Matrix.dotProduct (P - Q) (R - Q) = 0)
+(threesquare : EuclideanSpace ℝ (Fin 2) → EuclideanSpace ℝ (Fin 2) → EuclideanSpace ℝ (Fin 2) → Prop := fun P Q R ↦ dist Q P = dist Q R ∧ ⟪P - Q, R - Q⟫_ℝ = 0)
```
GeorgeTsoukalas, line 17: Could you clarify why using Matrix.dotProduct is error free? Also, looks like this is also an optParam, must have missed it earlier. Could you modify that?

eric-wieser: Nice catch, this is indeed an optParam!

eric-wieser: #236 finds two other stray optParams

eric-wieser: > Could you clarify why using Matrix.dotProduct is error free? 

I've expanded the PR description.

GeorgeTsoukalas (approved):

PR #234, opened 2024-10-21
Remove `open BigOperators` which does nothing

eric-wieser:
This used to be necessary, but the notation is available globally in recent mathlib versions.


eric-wieser:
The failure here is fixed by #237;

GeorgeTsoukalas (approved):

GeorgeTsoukalas:
Thanks for the contribution!

PR #238, opened 2024-10-22
Fix a few minor Lean misformalisations

ocfnash:

PR #239, opened 2024-10-23
Standardize on `:= sorry` not `:= by sorry`

eric-wieser:
_@GeorgeTsoukalas in https://github.com/trishullab/PutnamBench/pull/238#discussion_r1812821675_ said:

> Could you remove `by` here? The convention for the rest of the Lean formalizations is not to enter tactic mode by default.


GeorgeTsoukalas (approved):
Whoops! Looks like I missed these earlier. Thanks for the contribution Eric!

PR #243, opened 2024-11-01

ocfnash:


lean4/src/putnam_1984_a6.lean
```diff
@@ -19,6 +19,7 @@ theorem putnam_1984_a6
     (P_def : ∀ g p, P g p ↔
       if p = 0 then (∀ q > 0, ¬ Periodic g q) else IsLeast {q | 0 < q ∧ Periodic g q} p) :
     ∃ g : ℕ → ℕ,
-      (∀ᵉ (k > 0) (a : Fin k → ℕ) (ha : Injective a), f (∑ i, 5 ^ (a i)) = g (∑ i, a i)) ∧
+      (∀ᵉ (k > 0) (a : Fin k → ℕ) (ha : Injective a) (ha' : 0 < ∑ i, a i),
+        f (∑ i, 5 ^ (a i)) = g (∑ i, a i)) ∧
```
GeorgeTsoukalas, line 23: Doesn't leaving `g 0` unconstrained mess up claims about periodicity? I also think the former choice (that `g` is periodic for $s \geq 1$ is more aligned with the informal statement which requires exactly that assumption in (b).

ocfnash: Oooh you're right. I was about to say that all was well because the periodicity pinned `g 0` but the fact that part of the answer is about whether `g` was periodic means this indeed breaks the statement.

I have adjusted accordingly.

GeorgeTsoukalas (approved):
Looks good to me! As always, thanks for the contribution, Oliver!

PR #245, opened 2024-11-09
Fix some formalizations, provide few Lean rewrites.

GeorgeTsoukalas:


lean4/src/putnam_1987_a5.lean
```diff
@@ -13,21 +11,13 @@ Let $\vec{G}(x,y)=\left(\frac{-y}{x^2+4y^2},\frac{x}{x^2+4y^2},0\right)$. Prove
...
-: (∃ F : Fin 3 → ((Fin 3 → ℝ) → ℝ), Fprop1 F ∧ Fprop2 F ∧ Fprop3 F) ↔ putnam_1987_a5_solution :=
-sorry
+    (G : (Fin 2 → ℝ) → (Fin 3 → ℝ))
+    (G_def : ∀ x y, G ![x, y] = ![(-y / (x ^ 2 + 4 * y ^ 2)), (x / (x ^ 2 + 4 * y ^ 2)), 0]) :
+    (∃ F : (Fin 3 → ℝ) → (Fin 3 → ℝ),
+      ContDiffOn ℝ 1 F {v | v ≠ ![0,0,0]} ∧
+      (∀ x, x ≠ 0 →
+        (fderiv ℝ F x (Pi.single 1 1) 2 - fderiv ℝ F x (Pi.single 2 1) 1 = 0 ∧
+        fderiv ℝ F x (Pi.single 2 1) 0 - fderiv ℝ F x (Pi.single 0 1) 2 = 0 ∧
+        fderiv ℝ F x (Pi.single 0 1) 1 - fderiv ℝ F x (Pi.single 1 1) 0 = 0)) ∧
```
GeorgeTsoukalas, line 21: @eric-wieser Could you spare a quick moment to inspect the usage of `fderiv` here? Would be greatly appreciated!

eric-wieser: Gosh, this one is a monster. My preference here would be to add
```lean
    (curl : ((Fin 3 → ℝ) → (Fin 3 → ℝ)) → ((Fin 3 → ℝ) → (Fin 3 → ℝ)))
    (curl_def : ∀ f x, curl f x = ![
      fderiv ℝ f x (Pi.single 1 1) 2 - fderiv ℝ f x (Pi.single 2 1) 1,
      fderiv ℝ f x (Pi.single 2 1) 0 - fderiv ℝ f x (Pi.single 0 1) 2,
      fderiv ℝ f x (Pi.single 0 1) 1 - fderiv ℝ f x (Pi.single 1 1) 0])
```
and then use
```lean
        (∀ x, x ≠ 0 → curl F x = 0)
```
here

lean4/src/putnam_1986_a5.lean
```diff
@@ -1,22 +1,12 @@
...
-(hcontdiffx : contdiffx = (fun (func : (Fin n → ℝ) → ℝ) (i : Fin n) (x : Fin n → ℝ) => ContDiff ℝ 1 (fun xi : ℝ => func (xrepl x i xi))))
-(hfcontdiff1 : ∀ i : Fin n, ∀ j : Fin n, ∀ x : Fin n → ℝ, contdiffx (f i) j x)
-(hfcontdiff2 : ∀ i : Fin n, ∀ j1 j2 : Fin n, ∀ x : Fin n → ℝ, contdiffx (partderiv (f i) j1) j2 x)
-(hfc : ∃ c : Fin n → Fin n → ℝ, ∀ i j : Fin n, partderiv (f i) j - partderiv (f j) i = (fun x : Fin n → ℝ => c i j))
-: ∃ g : (Fin n → ℝ) → ℝ, ∀ i : Fin n, IsLinearMap ℝ (f i + partderiv g i) :=
-sorry
+  (n : ℕ) (hn : 1 ≤ n)
+  (f : Fin n → ((Fin n → ℝ) → ℝ))
+  (hf : ∀ i, ContDiff ℝ 2 (f i))
+  (hf' : ∀ i j : Fin n, ∃ C : ℝ, ∀ x : Fin n → ℝ, fderiv ℝ (f i) x (Pi.single j 1) - fderiv ℝ (f j) x (Pi.single i 1) = C)
```
eric-wieser, line 10: As a very minor nit, I would write this
```suggestion
  (C : Fin n → Fin n → ℝ)
  (hf' : ∀ i j : Fin n, ∀ x : Fin n → ℝ, fderiv ℝ (f i) x (Pi.single j 1) - fderiv ℝ (f j) x (Pi.single i 1) = C i j)
```
to match the $c_{ij}$ in the prose

lean4/src/putnam_1989_b6.lean
```diff
@@ -1,23 +1,19 @@
...
-sorry
+  (n : ℕ) (hn : 0 < n)
+  (S : (ℝ → ℝ) → (Fin (n + 2) → Icc (0 : ℝ) 1) → ℝ)
+  (hS : ∀ f x, S f x = if StrictMono x ∧ x 0 = 0 ∧ x (-1) = 1 then ∑ i in Icc 0 n, (x (i + 1) - x i) * f (x (i + 1)) else 0)
+  : ∃ P : Polynomial ℝ,
+    P.degree = n ∧
+    (∀ t ∈ Icc 0 1, P.eval t ∈ Icc 0 1) ∧
+    (∀ f : ℝ → ℝ,
+      f 1 = 0 ∧ ContinuousOn f (Icc 0 1) →
+      𝔼[(↑) ∘ (S f)] = ∫ t in (0)..1, (f t) * (P.eval t)) :=
```
eric-wieser, line 18: I think this would be better as
```suggestion
      𝔼[fun x => S f x] = ∫ t in (0)..1, f t * P.eval t) :=
```
which makes it clear where the `x` in the question has got to

lean4/src/putnam_1989_b6.lean
```diff
@@ -1,23 +1,19 @@
...
-(hfprop : fprop = (fun f : ℝ → ℝ => ContinuousOn f (Set.Icc 0 1) ∧ f 1 = 0))
-(hfxsum : fxsum = (fun (f : ℝ → ℝ) (x : Fin n → ℝ) => ∑ i in Finset.Icc 0 n, ((xext x) (i + 1) - (xext x) i) * f ((xext x) (i + 1))))
-(hfEV : fEV = (fun f : ℝ → ℝ => (∫ x in Sx, fxsum f x) / (∫ x in Sx, 1)))
-(npos : n ≥ 1)
-(hxext : ∀ x : Fin n → ℝ, (xext x) 0 = 0 ∧ (xext x) (n + 1) = 1 ∧ (∀ i : Fin n, (xext x) (i + 1) = x i))
-: ∃ P : Polynomial ℝ, P.degree = n ∧ (∀ t ∈ Set.Icc 0 1, 0 ≤ P.eval t ∧ P.eval t ≤ 1) ∧ (∀ f : ℝ → ℝ, fprop f → fEV f = (∫ t in Set.Ioo 0 1, f t * P.eval t)) :=
-sorry
+  (n : ℕ) (hn : 0 < n)
+  (S : (ℝ → ℝ) → (Fin (n + 2) → Icc (0 : ℝ) 1) → ℝ)
+  (hS : ∀ f x, S f x = if StrictMono x ∧ x 0 = 0 ∧ x (-1) = 1 then ∑ i in Icc 0 n, (x (i + 1) - x i) * f (x (i + 1)) else 0)
```
eric-wieser, line 12: This summation results in some pretty funky casting. I think it might be better to write `∑ i : Fin n.succ, (x i.succ - x i.castSucc) * f (x i.succ)`, though perhaps @ocfnash can give a second opinion when they're back from vacation.

lean4/src/putnam_1989_b6.lean
eric-wieser, line 18: Separately, I worry that this expected value is incorrect, as it is not conditioning on the fact that `x` lies within the appropriate region

GeorgeTsoukalas: I think that you're right. `x` does not lie in the appropriate region, and although the S f should be 0 in the extra space, it should still affect the expectation. What's a good workaround? I could write out an integral for the expectation explicitly over the right region.

lean4/src/putnam_1989_b6.lean
GeorgeTsoukalas: @ocfnash What are your thoughts on this?

lean4/src/putnam_1989_b6.lean
ocfnash: Yeah this is an `n + 2`-dimensional integral instead of an `n`-dimensional one. Indeed, the infoview confirms that the measure used for this expectation is  `(volume : Measure <| Fin (n + 2) → Icc (0 : ℝ) 1)`. Since `S f` is zero almost everywhere in this region (where we fail to have `x 0 = x (-1)`) I believe this expectation is zero.

lean4/src/putnam_1989_b6.lean
ocfnash: Sorry I'm so horrendously late replying here; I think I side with Eric.

lean4/src/putnam_1989_b6.lean
ocfnash: I have some time this morning so I'll see if I can write a formalisation which avoids the issues identified.

ocfnash: I think the following is correct:
```lean
theorem putnam_1989_b6
    (n : ℕ) [NeZero n]
    (I : (Fin n → ℝ) → Fin (n + 2) → ℝ)
    (I_def : ∀ x i, I x i = if i = 0 then 0 else if i = - 1 then 1 else x (i : ℕ).pred)
    (X : Set (Fin n → ℝ))
    (X_def : ∀ x, x ∈ X ↔ 0 < x 0 ∧ x (-1) < 1 ∧ ∀ i, i + 1 < n → x i < x (i + 1))
    (S : (ℝ → ℝ) → (Fin (n + 2) → ℝ) → ℝ)
    (S_def : ∀ f x, S f x = ∑ i in Finset.Iic n, (x (i + 1) - x i) * f (i + 1)) :
    ∃ P : Polynomial ℝ,
      P.degree = n ∧
      (∀ t ∈ Icc 0 1, P.eval t ∈ Icc 0 1) ∧
      (∀ f : ℝ → ℝ, f 1 = 0 → ContinuousOn f (Icc 0 1) →
        ∫ x, S f (I x) ∂ℙ[|X] = ∫ t in (0)..1, f t * P.eval t) :=
  sorry
```

GeorgeTsoukalas: I think this looks right to me as well - I will update the PR. Thank you for providing a fix!

PR #246, opened 2024-11-14
lean: remove incorrect type annotation in putnam_1996_a2.lean

eric-wieser:
These should be `EuclideanSpace ℝ (Fin 2)` not `Fin 2 → ℝ`, but if omitted lean works it out anyway.

This doesn't affect the correctness of the problem statement.


GeorgeTsoukalas:
Thanks!

PR #248, opened 2024-12-09
Add Putnam 2024 Lean files and some initial formalizations

GeorgeTsoukalas:
Currently 9 of 12 problems are given formalizations with the remaining three relying on geometry & probability theory which appear to be trickier to formalize. Solutions and informal statements have not been added because they have yet to be officially released.


GeorgeTsoukalas:
For folks curious about this PR: we are in discussion with another team to add some standardized formalizations from this year's Putnam.

PR #249, opened 2025-01-03
Bump Lean to version 4.14.0

GeorgeTsoukalas:
Bump Lean to version 4.14.0. Most changes to Lean files are because of `FiniteDimensional.finrank` now being switched to `Module.finrank`


GeorgeTsoukalas:
@eric-wieser It seems there is an issue in `check-docstrings` on line 98. Could you take a look? (Since I'm not familiar with using Lean for these purposes I don't want to make a syntactic fix which happens to change the semantics)

eric-wieser:
Is this intended to target Putnam2024 rather than main / master?

eric-wieser:
> @eric-wieser It seems there is an issue in `check-docstrings` on line 98. Could you take a look?

The fix is `import Lean.Util.SearchPath` at the top of the file.

eric-wieser:
I've made the fix in #250. Would you mind merging that first, so as to ensure a version of PutnamBench exists for v4.13.0 too?

lean4/src/putnam_1989_a6.lean
```diff
@@ -17,6 +17,8 @@ theorem putnam_1989_a6
 (F : Type*) [Field F] [Fintype F]
 (hF : Fintype.card F = 2)
 (α : PowerSeries F)
-(hα : ∀ n : ℕ, let bin := [1] ++ (digits 2 n) ++ [1]; PowerSeries.coeff F n α = ite (∀ i j : Fin bin.length, i < j → bin.get i = 1 → bin.get j = 1 → (∀ k, i < k → k < j → bin.get k = 0) → Even ((j : ℕ) - (i : ℕ) - 1)) 1 0)
```
eric-wieser, line 20: This can be fixed with `set_option synthInstance.maxSize $random_number`, which I am reporting now...

eric-wieser: https://leanprover.zulipchat.com/#narrow/channel/270676-lean4/topic/Success.20is.20not.20monotone.20in.20.60synthInstance.2EmaxSize.60/near/491823649

PR #253, opened 2025-01-05
Avoid using Mathlib.Vector

eric-wieser:
This isn't idiomatic for mathematics, and the name changes in 4.15.0. Avoiding it entirely means that no changes are required to bump to 4.15.0 (and so PutnamBench works on both versions)


GeorgeTsoukalas (approved):
Good to know - thanks for the contribution!

PR #255, opened 2025-01-10
Fix 2012 A4

ocfnash:
This was disprovable because of the `n > 2` condition in the goal.
For example if we take:
 * $q = 101$
 * $r = 1$
 * $A = [0, 100]$
 * $B = [1, 2]$
    
Then $S = \\{1, 2\\}$ and there is no three-or-more term arithmetic progression whose intersection with $A$ is $S$.

### DIFF
--- a/lean4/src/putnam_2012_a4.lean
+++ b/lean4/src/putnam_2012_a4.lean
@@ -6,13 +6,17 @@ open Matrix Function
 Let $q$ and $r$ be integers with $q>0$, and let $A$ and $B$ be intervals on the real line. Let $T$ be the set of all $b+mq$ where $b$ and $m$ are integers with $b$ in $B$, and let $S$ be the set of all integers $a$ in $A$ such that $ra$ is in $T$. Show that if the product of the lengths of $A$ and $B$ is less than $q$, then $S$ is the intersection of $A$ with some arithmetic progression.
 -/
 theorem putnam_2012_a4
-(q r : ℤ)
-(A B : Fin 2 → ℝ)
-(T : Set ℝ)
-(S : Set ℤ)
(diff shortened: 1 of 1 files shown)


lean4/src/putnam_2012_a4.lean
```diff
@@ -6,13 +6,17 @@ open Matrix Function
...
+    (q r : ℤ)
+    (A B : Fin 2 → ℝ)
+    (T : Set ℤ)
+    (S : Set ℤ)
+    (qpos : q > 0)
+    (ABlt : A 0 < A 1 ∧ B 0 < B 1)
+    (hT : T = {x : ℤ | ∃ b m : ℤ, (b : ℝ) ∈ Set.Icc (B 0) (B 1) ∧ x = b + m * q})
+    (hS : S = {a : ℤ | (a : ℝ) ∈ Set.Icc (A 0) (A 1) ∧ r * a ∈ T})
+    (prod_lt : (A 1 - A 0) * (B 1 - B 0) < q) :
+    IsFiniteAP {x | x ∈ S ∧ (x : ℝ) ∈ Set.Icc (A 0) (A 1)} :=
```
GeorgeTsoukalas, line 21: This final goal line states that the A \cap S is an arithmetic progression, but the problem statement says S is A \cap X where X is a set of points of an (finite) AP.  The latter implies the former but my feeling is that the goal as stated does not reflect the informal statement, what do you think?

GeorgeTsoukalas: On second thought I think the difference is minimal enough to not warrant a change, and the other statement may include some additional casting which could make the problem unnecessarily harder to prove.

PR #256, opened 2025-01-10
fix putnam_2022_a6

mo271:
Here the way the length of the interval was measured should be switched, so it has always positive length.

Drive-by: remove a superfluous `Finset`


GeorgeTsoukalas:
Good catch! Thanks for the contribution!

PR #266, opened 2025-02-25
Avoid Group.closure

mo271:


mo271:
No, `AddGroup.closure` is not used anywhere.

PR #268, opened 2025-02-28
Set won reflexive to Boolean false

sorgfresser:
Currently, won is defined with co-domain Bool. The reflexive hypothesis assures `won i i = False` using the Prop `False`. Instead, we should use Bool here.


GeorgeTsoukalas (approved):
Thanks for the contribution!

PR #270, opened 2025-02-28
feat: Add formalisation of Putnam 1968 B1

Paul-Lez:


lean4/src/putnam_1968_b1.lean
```diff
@@ -0,0 +1,24 @@
+import Mathlib
+
+open MeasureTheory
+open scoped ProbabilityTheory
+
+def putnam_1968_b1_solution : ℝ → ℝ → ℝ → ℝ := sorry
+
```
GeorgeTsoukalas, line 7: Can you include the solution as well as a comment on line 7? This is the current convention for the Lean problems.

GeorgeTsoukalas (approved):
Looks good. Thank you for the contribution, Paul!

PR #276, opened 2025-03-17
feat: Add formalisation of putnam_1964_b4

Paul-Lez:


lean4/src/putnam_1964_b4.lean
```diff
@@ -0,0 +1,32 @@
...
+$n$ great circles on the sphere are in general position (in other words at most two circles pass through any two points on the sphere). How many regions do they divide the sphere into?
+-/
+theorem putnam_1964_b4
+    {n : ℕ} (hn : 0 < n)
+    -- `C` is a collection of `n` great circles on the sphere, i.e a collection of sets
+    (C : Fin n → Set (EuclideanSpace ℝ (Fin 3)))
+    --together with a collection of `n` normal vectors `v`
+    (v : Fin n → EuclideanSpace ℝ (Fin 3))
+    -- A great circle `C` is the intersection of a plane `{x | x · v = 0}` with `S`
+    (hv : ∀ i, C i = Metric.sphere 0 1 ∩ {x : EuclideanSpace ℝ (Fin 3) | ⟪v i, x⟫_ℝ = 0 })
```
GeorgeTsoukalas, line 19: Think this requires a restriction v i is nonzero right? Otherwise the circle `C i` could be `Metric.sphere 0 1`

Paul-Lez: Nice catch! Fixed.

lean4/src/putnam_1964_b4.lean
```diff
@@ -0,0 +1,34 @@
...
+    (hT₂ : ∀ᵉ (x ∈ Metric.sphere 0 1) (y ∈ Metric.sphere 0 1),
+      (Finset.univ.filter (fun i => {x, y} ⊆ (C i))).card ≤ 2)
+    -- The predicate that a set `R ⊆ ℝ³` is a region bounded by the collection of circles
+    (IsRegion : Set (EuclideanSpace ℝ (Fin 3)) → Prop)
+    -- We can determine the points `x` in such a region by looking at the sign of their dot product
+    -- with `n i` for `i : Fin n`
+    (IsRegion_def : ∀ R, IsRegion R ↔
+      R ⊆ Metric.sphere 0 1 ∧ R.Nonempty ∧ ∃ sign : Fin n → SignType,
+      (∀ i, sign i ≠ 0) ∧
+      R = {x : EuclideanSpace ℝ (Fin 3) | ∀ i, signHom ⟪v i, x⟫_ℝ = sign i}) :
```
GeorgeTsoukalas, line 32: I think no set region will ever qualify for `IsRegion` right? Because R is retricted to be a subset of Metric.sphere 0 1 so the check R = {x : EuclideanSpace R (Fin 3) | sign <v_i, x> = sign i} will never be true?

Paul-Lez: Right, the set is missing an intersection with the unit sphere (added it now)

GeorgeTsoukalas (approved):
Looks good to me, thanks for the contribution, Paul! Looking forward to the next one :)

PR #281, opened 2025-04-01
feat: Add formalisation of Putnam 1969 B4

Paul-Lez:


lean4/src/putnam_1969_b4.lean
```diff
@@ -0,0 +1,20 @@
...
+
+/--
+$Γ$ is a plane curve of length 1. Show that we can find a closed rectangle of area 1/4 which covers $Γ$.
+-/
+theorem putnam_1969_b4
+    (Γ : ℝ → EuclideanSpace ℝ (Fin 2))
+    --Note: the problem doesn't say what regularity conditions we should impose on `Γ` - hopefully continuity is enough.
+    (Γ_cts : ContinuousOn Γ (Set.Icc 0 1))
+    (hΓ : eVariationOn Γ (Set.Icc 0 1) = 1) :
+    letI : Module.Oriented ℝ (EuclideanSpace ℝ (Fin 2)) (Fin 2) := ⟨Basis.orientation <| Pi.basisFun _ _⟩
```
GeorgeTsoukalas, line 13: Not very familiar with this part, is this so that you can use angles?

lean4/src/putnam_1969_b4.lean
```diff
@@ -0,0 +1,20 @@
+import Mathlib
+
+open scoped Real EuclideanGeometry
+
+/--
+$Γ$ is a plane curve of length 1. Show that we can find a closed rectangle of area 1/4 which covers $Γ$.
+-/
+theorem putnam_1969_b4
+    (Γ : ℝ → EuclideanSpace ℝ (Fin 2))
+    --Note: the problem doesn't say what regularity conditions we should impose on `Γ` - hopefully continuity is enough.
```
GeorgeTsoukalas, line 10: Looking at the solution I believe that it is enough. In general a continuous curve is not rectifiable but with the condition that eVariation = 1 I think it should force the correct behavior anyways.

lean4/src/putnam_1969_b4.lean
Paul-Lez: Yes exactly: to use `EuclideanGeoemtry.oangle` you need `[Fact (Module.finrank ℝ V = 2)] [Module.Oriented ℝ V (Fin 2)]` (where `V` is your ambiant space).

GeorgeTsoukalas (approved):
Looks good to me, thanks for the contribution!

PR #288, opened 2025-06-03
feat: Add formalisation of Putnam 2022 a4

Paul-Lez:
I still have a few sanity checks I'd like to do on the definition of `k` - maybe worth postponing the review until then:)


GeorgeTsoukalas (approved):
Looks good to me, thanks for the contribution Paul!

PR #290, opened 2025-07-03
lean: fix deprecation warnings

eric-wieser:
The big operator notation changed upstream, and `Complex.abs` is now deprecated in favor of the more general `norm`.


GeorgeTsoukalas (approved):
Thanks for the changes!

PR #303, opened 2025-08-28
Remove `Nat.cast` from `Fin`, in preparation for a Lean upgrade

eric-wieser:
The new behavior can be emulated with `attribute [-instance] Fin.instCommRing Fin.instNatCast Fin.instAddMonoidWithOne`.

While the old behavior can be kept with `open scoped Fin.CommRing`, the spelling without is usually clearer anyway.


GeorgeTsoukalas (approved):
Looks good to me, thanks Eric!

PR #304, opened 2025-08-29
Update to newer lean versions

eric-wieser:
This PR will contain a commit for every major version of Lean; but I will wait for CI to complete on each before pushing the next.


GeorgeTsoukalas:
Thank you so much, Eric!

PR #306, opened 2025-11-07
lean: change the solution for putnam_1968_b1 to an abbrev

eric-wieser:
Otherwise the script to inject solutions doesn't run.


GeorgeTsoukalas (approved):
Thank you Eric!

PR #310, opened 2025-11-26
Fix JSON syntax for GPT-5 entry in results

winger:


GeorgeTsoukalas:
Woah! Good catch. Must have been half asleep when I submitted #309.

PR #355, opened 2026-09-14
Fix solution predicate for Putnam 1974 B1

### LINKED ISSUE #347: `putnam_1974_b1` appears refutable
## Summary

The Lean formalization of `putnam_1974_b1` appears to remain false in the official PutnamBench source, despite the historical fix that changed the regular-pentagon answer predicate to require `B > 0`.

The adjacent official answer comment is:

```lean
fun p ↦ ∃ᵉ (B > 0) (o : Equiv.Perm (Fin 5)),
  ∀ i, dist (p (o i)) (p (o (i + 1))) = B
```

## Observed issue

The predicate appears to permit repeated non-adjacent points.  A concrete unit-circle configuration of the shape `[A, B, C, A, B]`, where `A`, `B`, and `C` are vertices of an equilateral triangle on the unit circle, satisfies all five cyclic equal-distance constraints with the positive value `B = √3`. However, its pairwise-distance sum is strictly below that of another valid unit-circle configuration.  

## Reproduction

To reduce contamination, please contact me for a complete Lean 4.27 counterexample file if needed.

danbq2289:
Fixes #347.

The existing solution predicate only requires the five cyclic consecutive distances to be equal and positive. Since p is not required to be injective, this admits degenerate configurations with repeated points, e.g. [A, B, C, A, B] for an equilateral triangle on the unit circle.

This adds Function.Injective p to exclude such repeated-point configurations while preserving the existing one-line solution format.

This fix was previously noted by Near AI; I independently encountered the issue while validating our Midas Prover submission and independently verified in Lean that the corrected predicate proves the theorem.


GeorgeTsoukalas:
Awesome, thanks!!

